PR #144, opened 2025-01-22
chore: update CONTRIBUTING.md

0xRAG:
## Summary

This PR updates the contributing guide. It adds a high-level generic guide, as well as two language specific guides – one for Python and one for TypeScript


cb-heimdall:
<table>
<tr>
<th>Requirement</th>
<th>Status</th>
<th>More Info</th>
</tr>
<tr>
<td>Reviews</td>
<td>
🟡
<code>0/1</code>
</td>
<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>
</td>
<td>1</td>
</tr>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Max</strong></td>
<td><strong>0</strong></td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>
From <code>CODEOWNERS</code>
</td>
<td>0</td>
</tr>
<tr>
<td>Global minimum</td>
<td>0</td>
</tr>
<tr>
<td><strong>Max</strong></td>
<td>
<strong>
1
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>1</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>1</strong></td>
</tr>
</table>
</details>
</td>
</tr>
</table>

John-peterson-coinbase:
I question if we should have the steps interlaced with Python and Typescript. I imagine a developer will follow the complete contribution guide top to bottom in 1 language (they may then go to the other language, but it is not required). There is a high level of focus required to go through the full doc and only parse out the Python sections for example without missing any.

It might be best to split into sub-documents for python and typescript. The top level doc can give higher level advice such as how to discuss a feature request / bug via GH Issue, PR review triage and SLA, etc.. 

cc: @murrlincoln for thoughts

murrlincoln:
> I question if we should have the steps interlaced with Python and Typescript. I imagine a developer will follow the complete contribution guide top to bottom in 1 language (they may then go to the other language, but it is not required). There is a high level of focus required to go through the full doc and only parse out the Python sections for example without missing any.
> 
> It might be best to split into sub-documents for python and typescript. The top level doc can give higher level advice such as how to discuss a feature request / bug via GH Issue, PR review triage and SLA, etc..
> 
> cc: @murrlincoln for thoughts

I agree this would make the most sense - if we put this in the main docs (I'd like to) then we can put it hidden beneath tabs but for this style of documentation it seems to make sense to separate them both

README.md
```diff
@@ -82,7 +82,7 @@ See [Twitter Langchain](./twitter-langchain/README.md) to get started!
 ## Contributing
 
 AgentKit welcomes community contributions.
-See [CONTRIBUTING.md](CONTRIBUTING.md) for more information.
+See [CONTRIBUTING.md](./.github/CONTRIBUTING.md) for more information.
```
murrlincoln, line 85: Will defer to you, but does putting the contribution file inside the Github folder hide it from people coming to the repo for the first time? Would it make sense to keep CONTRIBUTING.md but inside of it simply link to the new guides?

.github/CONTRIBUTING-TYPESCRIPT.md
```diff
@@ -0,0 +1,83 @@
...
+
+## Adding an Agentic Action
+
+One of the most common ways to contribute to AgentKit is by adding a new agentic action. Here are the high level steps:
+
+**Note: It is recommended to point your editor to the specific package that you are working in. This will help ensure things like package imports are working correctly**
+
+1. Create a new file in `cdp-agentkit-core/typescript/src/actions/cdp`
+2. Implement your new action inside your newly created file
+    - For an example of an action, see [mint_nft.py](./cdp-agentkit-core/typescript/src/actions/cdp/mint_nft.ts)
```
murrlincoln, line 33: Could we add some more detail to both of these files or a link out to our docs about how these actions are constructed? could also mean just commenting this specific mint NFT example with an explanation of the prompt, input schema, and function

Would be great to define what's needed in the prompt to make a good call, and that it should reduce reliance on user inputs to facilitate autonomous agent interactions (so we don't have the morpho-style issue where a user still needs to grab a vault externally)

.github/CONTRIBUTING-TYPESCRIPT.md
```diff
@@ -0,0 +1,83 @@
...
+A good way to test new actions locally is by using the chatbot example in `cdp-langchain`. This requires a couple API keys:
+
+- Get a Coinbase Developer Platform API Key from the [CDP Portal](https://portal.cdp.coinbase.com/access/api)
+- Get an [OpenAI API Key](https://platform.openai.com/docs/quickstart#create-and-export-an-api-key)
+
+The flow is:
+
+1. Make your change as described in the [Adding an Agentic Action](#adding-an-agentic-action) section
+2. Re-build the package by running `npm run build` (either from root, or from the package folder will do)
+3. In `cdp-langchain/examples/chatbot-typescript`, run `npm run start`
```
murrlincoln, line 53: do we not need to point the package to the local version like we do in Python?

.github/CONTRIBUTING-TYPESCRIPT.md
```diff
@@ -0,0 +1,83 @@
+# TypeScript Development Guide
+
+This guide covers TypeScript-specific setup and development for AgentKit.
+
+## Contents
+
```
murrlincoln, line 6: Thoughts about putting a context section here or some sort of link out to CDP SDK for those using contract invocations? Want to prevent issues where someone uses an API or something that is an unnecessary dependency/requires each user to provision API keys (where possible. If there's no other option that's fine but we can't include it as a default tool)

.github/CONTRIBUTING-TYPESCRIPT.md
```diff
@@ -0,0 +1,83 @@
...
+- [Development Setup](#development-setup)
```
murrlincoln, line 7: May not be appropriate for this pass, but at some point it would be great to define how we want contributors to integrate AgentKit into new frameworks. Some of the prizes for the SF AI hackathon reward for this.

Main goal being to prevent a framework user from having to do anything but import the latest package for agentkit to get all of the most recent features and capabilities

README.md
0xRAG: I don't think so, since we're still linking to it here from the README. It's cleaner IMO to put all the GH-related files in the `.github` folder

.github/CONTRIBUTING-TYPESCRIPT.md
0xRAG: > Could we add some more detail to both of these files or a link out to our docs about how these actions are constructed?

For sure, we actually had this already but it was mostly repeating what's already explained via the code and comments in the code, so I decided to omit here. Happy to pull it back in – I think it makes sense to have its own doc since it will be mostly repeated between the Python and TypeScript guides

0xRAG: > Would be great to define what's needed in the prompt to make a good call, and that it should reduce reliance on user inputs to facilitate autonomous agent interactions

Great point

.github/CONTRIBUTING-TYPESCRIPT.md
0xRAG: Nope, we will not need to after a change I am making to the monorepo config

John-peterson-coinbase (approved):
Let's continue to iterate here. Merging so we can receive feedback

PR #155, opened 2025-01-23
chore: simplify internal package dependencies

0xRAG:
### What changed? Why?

Uses Turbo's [Internal Packages](https://turbo.build/repo/docs/core-concepts/internal-packages) so that we don't have to manually bump dependency versions in packages that depend on each other.


cdp-langchain/typescript/package.json
```diff
@@ -38,8 +38,7 @@
     "langchain"
   ],
   "dependencies": {
-    "@coinbase/cdp-agentkit-core": "^0.0.13",
-    "@coinbase/coinbase-sdk": "^0.14.1",
```
0xRAG, line 42: This is listed as a peerDependency, it does not need to be included here.

PR #181, opened 2025-01-30
chore: CDP, Farcaster, ERC721 action providers

John-peterson-coinbase:
### What changed? Why?
Added action providers for CDP, Farcaster and ERC721 actions

### DIFF
--- a/cdp-agentkit-core/typescript/src/action_providers/farcaster/farcasterActionProvider.ts
+++ b/cdp-agentkit-core/typescript/src/action_providers/farcaster/farcasterActionProvider.ts
@@ -1,0 +1,155 @@
+import { z } from "zod";
+import { ActionProvider } from "../action_provider";
+import { Network } from "../../wallet_providers";
+import { CreateAction } from "../action_decorator";
+import { AccountDetailsSchema, PostCastSchema } from "./schemas";
+
+/**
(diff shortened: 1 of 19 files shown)


cdp-agentkit-core/typescript/src/action_providers/farcaster/farcasterActionProvider.ts
```diff
@@ -0,0 +1,155 @@
...
+  supportsNetwork = (_: Network) => true;
+
+  /**
+   * Retrieves agent's Farcaster account details.
+   *
+   * @param _ The input arguments for the action.
+   * @returns A message containing account details for the agent's Farcaster account.
+   */
+  @CreateAction({
+    name: "account_details",
```
0xRAG, line 78: should the action names be camelCased too?

0xRAG (approved):
Looks like we're missing a README.md in erc721 as well, otherwise LGTM

cdp-agentkit-core/typescript/src/action_providers/farcaster/farcasterActionProvider.ts
John-peterson-coinbase: I would prefer for those to be the same across languages. Langchain does this as well

PR #214, opened 2025-02-01
feat: support custom action provider

0xRAG:
### What changed? Why?
Adds ability for devs to easily add a custom action provider:

```ts
    const custom = customActionProvider<EvmWalletProvider>({
      name: "custom_action",
      description: "This action does XYZ cool thing",
      schema: z.object({
        arg1: z.string(),
        arg2: z.string(),
      }),
      invoke: async (walletProvider, args: any) => {
        const { arg1, arg2 } = args;
        const address = walletProvider.getAddress();
        return `Invoked with ${arg1} and ${arg2} from ${address}`;
      },
    });

    const agentKit = await AgentKit.from({
      walletProvider,
      actionProviders: [custom],
    });
```

### DIFF
--- a/cdp-agentkit-core/typescript/src/action-providers/customActionProvider.ts
+++ b/cdp-agentkit-core/typescript/src/action-providers/customActionProvider.ts
@@ -1,0 +1,89 @@
+/* eslint-disable @typescript-eslint/no-explicit-any */
+
+import { z } from "zod";
+import { CreateAction } from "./actionDecorator";
+import { ActionProvider } from "./actionProvider";
+import { Network } from "../network";
+import { WalletProvider } from "../wallet-providers";
+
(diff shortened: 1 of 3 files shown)


cdp-agentkit-core/typescript/src/action-providers/actionDecorator.ts
```diff
@@ -120,7 +121,7 @@ export function CreateAction(params: CreateActionDecoratorParams) {
       Reflect.getMetadata(ACTION_DECORATOR_KEY, target.constructor) || new Map();
 
     const metaData: ActionMetadata = {
-      name: prefixedActionName,
+      name: originalName,
```
John-peterson-coinbase, line 124: why are we setting `name` to `originalName` instead of `prefixedActionName`

John-peterson-coinbase (approved):

PR #277, opened 2025-02-06
Enable source maps during the build process to simplify debugging.

irux:
### What changed? Why?

I just changed the source maps when building the library

#### Qualified Impact
Currently, the library is quite new and serves as a framework for building new tools. However, its documentation is still limited, making it difficult to understand its inner workings without attaching a debugger. Adding source maps will make debugging much easier and more convenient for users who install the package and begin to use the library.

### DIFF
--- a/tsconfig.base.json
+++ b/tsconfig.base.json
@@ -8,6 +8,7 @@
     "resolveJsonModule": true,
     "esModuleInterop": true,
     "moduleResolution": "node",
+    "sourceMap": true,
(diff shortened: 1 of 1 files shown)


0xRAG (approved):

0xRAG:
Hey @irux thanks for this. FYI you will need to enable commit signing in order to merge. See: https://docs.github.com/en/authentication/managing-commit-signature-verification/adding-a-gpg-key-to-your-github-account

PR #360, opened 2025-02-12
chore: add network mappings

0xRAG:
### What changed?
- [ ] Documentation
- [ ] Bug fix
- [ ] New Action
- [ ] New Action Provider
- [x] Other

### Why was this change implemented?
Adds network mappings to easily map from network_id to chain_id and chain data.

Testing locally with chatbot:
```
Prompt: what network are you on

-------------------
Wallet Details:
- Provider: cdp_wallet_provider
- Address: 0x234d74b45c3A942E425363C9C491C3C3B1A80986
- Network:
  * Protocol Family: evm
  * Network ID: base-sepolia
  * Chain ID: 84532
- Native Balance: 0
-------------------
I am on the Base Sepolia network.
-------------------
```

### DIFF
--- a/python/coinbase-agentkit/coinbase_agentkit/network/chain_definitions.py
+++ b/python/coinbase-agentkit/coinbase_agentkit/network/chain_definitions.py
@@ -1,0 +1,359 @@
+from pydantic import BaseModel
+
+
+class NativeCurrency(BaseModel):
+    """Represents the native currency for a blockchain network."""
+
+    name: str
+    symbol: str
+    decimals: int
+
+
+class RpcUrls(BaseModel):
+    """Represents the RPC URLs for a blockchain network."""
+
+    http: list[str]
+
+
+class BlockExplorer(BaseModel):
+    """Represents a block explorer for a blockchain network."""
+
+    name: str
+    url: str
+    api_url: str
+
+
+class Contract(BaseModel):
+    """Represents a contract on a blockchain network."""
+
+    address: str
+    block_created: int | None = None
+
+
+class Chain(BaseModel):
+    """Represents a blockchain network."""
+
+    id: int
+    name: str
+    network: str | None = None
+    native_currency: NativeCurrency
+    rpc_urls: dict[str, RpcUrls]
+    block_explorers: dict[str, BlockExplorer]
+    contracts: dict[str, Contract]
+    testnet: bool | None = False
+
+
+# Convert existing dictionaries to Chain instances
+mainnet = Chain(
+    id=1,
+    name="Ethereum",
+    native_currency={"name": "Ether", "symbol": "ETH", "decimals": 18},
+    rpc_urls={
+        "default": {
+            "http": ["https://eth.merkle.io"],
+        },
+    },
+    block_explorers={
+        "default": {
+            "name": "Etherscan",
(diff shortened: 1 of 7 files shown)


python/coinbase-agentkit/coinbase_agentkit/wallet_providers/eth_account_wallet_provider.py
```diff
@@ -9,16 +9,15 @@
...
 from .evm_wallet_provider import EvmWalletProvider
 
 
 class EthAccountWalletProviderConfig(BaseModel):
     """Configuration for EthAccountWalletProvider."""
 
     private_key: str
-    rpc_url: str
-    chain_id: int
+    network_id: str
```
John-peterson-coinbase, line 20: can we update this config to take an eth account instead of a private key so it is more flexible like the viem wallet provider?

python/coinbase-agentkit/coinbase_agentkit/network/chain_definitions.py
John-peterson-coinbase, line 1: Can we create a pydantic base model derived class for chain so these are validated and not just dictionaries?

python/coinbase-agentkit/coinbase_agentkit/wallet_providers/eth_account_wallet_provider.py
```diff
@@ -1,49 +1,56 @@
...
-    private_key: str
-    rpc_url: str
-    chain_id: int
+    account: LocalAccount
+    network_id: str
+
+    class Config:
+        """Configuration for EthAccountWalletProvider."""
+
+        arbitrary_types_allowed = True
```
0xRAG, line 25: Had to add this to silence a Pydantic error due to the addition of `LocalAccount` here

python/coinbase-agentkit/coinbase_agentkit/wallet_providers/eth_account_wallet_provider.py
```diff
@@ -1,49 +1,56 @@
...
 
 
 class EthAccountWalletProviderConfig(BaseModel):
     """Configuration for EthAccountWalletProvider."""
 
-    private_key: str
-    rpc_url: str
-    chain_id: int
+    account: LocalAccount
+    network_id: str
```
John-peterson-coinbase, line 20: I don't think `network_id` is necessary. Can we remove the `EthAccountWalletproviderConfig` object entirely and just pass the `Account` to the ctor. It also doesn't necessarily have to be a local account @0xRAG

0xRAG: sure, but should we maybe keep the config object in case the config evolves over time and we do need to end up adding more props? or do you not foresee that happening at all

John-peterson-coinbase: yes makes sense. lets just remove network ID

0xRAG: Also, isn't `network_id` necessary? To know which RPC URL to use etc

John-peterson-coinbase (approved):

PR #469, opened 2025-02-25
chore: introduce changesets

0xRAG:
### What changed?

This PR introduces [changesets](https://github.com/changesets/changesets) to manage the AgentKit.js changelog. The `.changesets` folder was generated with the changesets CLI command: `npx @changesets/cli`. The following config options were modified:
* [access](https://github.com/changesets/changesets/blob/main/docs/config-file-options.md#access-restricted--public): changed to `public` since we are publishing public packages
* [ignore](https://github.com/changesets/changesets/blob/main/docs/config-file-options.md#ignore-array-of-packages): a glob to ignore the examples, since these are not published. Note the docs say this should be handled via the `private: true` field in the package.json; however in practice these packages were still being surfaced by changesets
* [changelog](https://github.com/changesets/changesets/blob/main/docs/config-file-options.md#changelog-false-or-a-path): using `@changesets/changelog-github` for friendly and useful attribution (auto-tagging relevant PRs and a thank you message to contributors)

### Why?

Contributors can add independent changelog entries without having to worry about or deal with merge conflicts.

Maintainers can generate full changelogs and bump package versions with one command.

The intended contribution workflow is:
* Developer makes a change to some code in a package
* Developer runs `npx @changesets/cli` and follows the prompts to write a changeset
* Developer checks in generated changeset file with their PR
* Maintainer reviews the generated changeset file for version type and change content. Version type should follow [semver](https://semver.org/), and change content should be in past tense (e.g. "Added new feature", "Fixed the bug")

The release workflow is:
* Maintainer runs `npx @changesets/cli version` which will consume changelog entries in `.changesets` and append to the `CHANGELOG.md`, as well as bump package versions
* Maintainer opens PR with the result from above
* Once merged, packages are released as they are today, using existing GitHub Actions

This workflow can be further streamlined once we're comfortable with changesets, where we can have the changesets GitHub Action do the [versioning and publishing](https://github.com/changesets/action?tab=readme-ov-file#with-publishing) for us.

### How this changes the CHANGELOG.md format

As an example of a default `changesets`-formatted changelog, see the [@changesets/cli changelog](https://github.com/changesets/changesets/blob/main/packages/cli/CHANGELOG.md). Notably, changes are grouped by version type (i.e. `minor`, `patch`), instead of change type (`feature`, `bugfix`).

It's possible to [write custom formatting functions](https://github.com/changesets/changesets/blob/main/docs/modifying-changelog-format.md#writing-changelog-formatting-functions), but this introduces maintenance overhead and could be error-prone, so I would recommend we adopt the default strategy from `changesets`.

### Next steps
* Introduce [changesets bot](https://github.com/apps/changeset-bot) to automatically request contributors add a changeset
* Introduce [changesets GitHub Action](https://github.com/changesets/action) to handle versioning & publishing


CONTRIBUTING-PYTHON.md
```diff
@@ -23,8 +24,9 @@ poetry --version
 ```
 
 If the versions are not correct or you don't have Python or Poetry installed, download and follow their setup instructions:
-* Python: install with [pyenv](https://github.com/pyenv/pyenv)
-* Poetry: follow the official [Poetry installation instructions](https://python-poetry.org/docs/#installation)
+
```
0xRAG, line 27: These changes, through to the addition of the new Changelog section at the bottom, are all formatting fixes

CONTRIBUTING-PYTHON.md
```diff
@@ -221,3 +227,7 @@ make lint
 # Fix linting issues
 make lint-fix
 ```
+
+## Changelog
+
+Currently the CHANGELOG.md is managed manually. Please include a changelog entry in the appropriate package's CHANGELOG.md file when you add a new feature or fix a bug.
```
0xRAG, line 233: There will be a fast-follow to introduce a Python-specific changelog management tool in a separate PR. That PR will update these instructions

CarsonRoscoe (commented):
Couple questions:

1. Do we have control over the text in this changelog CLI? i.e. the space/enter error message, if it were up to me, I would inform the user "Select each option with space, and press enter when you're done"

Screenshot 2025-02-26 at 8 21 45 AM


2. Do we have control over the names of the changelog files? Mine generated as `crazy-days-bet.md`

3. If we don't have control over the changelog file names, can I assume at least they are consumed and deleted when the maintainer runs the command to that combines the changelogs before release?

0xRAG:
Thanks for the questions @CarsonRoscoe!

On the first, I didn't see anything about that in the docs, but I'll check again. I did already include a blurb about this in the TypeScript Contributing guide, but it would be nice to have that tip directly in the CLI.

On the second, yes the auto-generated filenames are consumed when running `npx @changesets/cli version`!

0xRAG:
Also on naming, see: https://github.com/changesets/changesets/blob/main/docs/common-questions.md#each-changeset-is-its-own-file

> We use random human readable names by default for these files to avoid collisions when generating them, but there's no harm that will come from renaming them.

I do like this approach as the contributor doesn't need to make a decision on what the filename should be. Once the changelog is ready to be updated, all changesets will be consumed and deleted.

CarsonRoscoe (approved):
Thanks for the clarifications,

LGTM

PR #566, opened 2025-03-14
feat: add across action provider (ts)

phdargen:
## Description

Adds new action provider to perform cross-chain interactions via the Across protocol. 
Across SDK is added as new dependency.

Implements:
- bridge_token action to bridge from connected chain to another evm chain
-> supports native ETH and ERC-20 transfers (as long as supported by Across)
- check_deposit_status action to check fill status on destination chain (mainnets only)

## Tests

- [x] Added unit tests
- [x] Tested prompts

```
Chatbot: typescript/examples/langchain-cdp-chatbot/chatbot.ts
Network: Sepolia
Setup: Fauceted with 1 USDC and ETH

Prompt: bridge to chain 84532, amount 0.001 ETH, receipienct 0xb7f17bF94DBa50077be037CDc279a868f82D86C2

-------------------
Successfully deposited tokens:
- From: Chain 11155111 (Sepolia)
- To: Chain 84532 (Base Sepolia)
- Token: ETH (0xfFf9976782d46CC05630D1f6eBAb18b2324d6B14)
- Input Amount: 0.001 ETH
- Output Amount: 0.000990746891309023 ETH
- Recipient: 0xb7f17bF94DBa50077be037CDc279a868f82D86C2
- Transaction Hash for deposit: 0x0f2c064c36991d9af1f0d81d347d5e16f61b6ea1302f430c1a57b07bec743cea
- Deposit ID: 1002457
-------------------

Prompt: bridge to chain 84532, amount 0.1 usdc, receipienct 0xb7f17bF94DBa50077be037CDc279a868f82D86C2, max slippage 3%
The bridging transaction was successful! Here are the details:

- **From Chain:** 11155111 (Sepolia)
- **To Chain:** 84532 (Base Sepolia)
- **Token:** USDC
- **Input Amount:** 0.1 USDC
- **Output Amount:** 0.098182 USDC
- **Recipient:** 0xb7f17bF94DBa50077be037CDc279a868f82D86C2
- **Transaction Hash for Approval:** 0x33575547e8031d3223e63dbc2e82fbc307b1391bd9c6379c8a91952b47024e92
- **Transaction Hash for Deposit:** 0x797f7d27e820f1832f994b6a6218118803fbc500f18caa70179811b4be21b2aa

Prompt: check status of deposit 1349975 with originChainId 137
-------------------
{
  "status": "filled",
  "depositTxInfo": {
    "txHash": "0x784f3cf234ffc960d087c5c02b166d838f7a170b337a349a49b54be837fd8152",
    "chainId": 137,
    "chainName": "Polygon"
  },
  "fillTxInfo": {
    "txHash": "0x788835d45d1ad5bc339990b23d2e09756ca1b4c98a6246be3505fb1baaf573e6",
    "chainId": 8453,
    "chainName": "Base"
  },
  "depositRefundTxInfo": null
}
-------------------
The status of the deposit with ID **1349975** and originChainId **137** is **filled**. Here are the transaction details:

- **Deposit Transaction Hash (Polygon)**: `0x784f3cf234ffc960d087c5c02b166d838f7a170b337a349a49b54be837fd8152`
  
- **Fill Transaction Hash (Base)**: `0x788835d45d1ad5bc339990b23d2e09756ca1b4c98a6246be3505fb1baaf573e6`

No refund transaction is related to this deposit.
-------------------

```

## Checklist

- [x] Added documentation to all relevant README.md files
- [x] Added a changelog entry


typescript/agentkit/src/action-providers/across/acrossActionProvider.ts
```diff
@@ -0,0 +1,285 @@
...
+          address: inputToken,
+          abi: ERC20_ABI,
+          functionName: "approve",
+          args: [quote.deposit.spokePoolAddress, approvalAmount],
+        });
+        await this.#publicClient.waitForTransactionReceipt({ hash: approvalTxHash });
+      }
+
+      // Simulate the deposit transaction
+      const { request } = await acrossClient.simulateDepositTx({
```
KanishkKhurana, line 247: Diving deep into the logic behind this, are we simulating txn and executing the deposit fn manually because there's no way to trigger a wallet extension ? 
I ask because we have the executeQuote function which will be able to wrap complexities around txn simulation and executing the deposit txn. 

overall this is a stellar job. thank you for putting in the effort here!

KanishkKhurana (commented):
left one comment around manual txn simulation and executing the deposit txn. 

overall, really good job from an Across Protocol integration POV

typescript/agentkit/src/action-providers/across/acrossActionProvider.ts
phdargen: Hi @KanishkKhurana , thank you for the feedback!

I initially used executeQuote which would indeed simplify the process. This works fine until progress.step === "deposit" is succesful but then throws errors in the progress.step === "fill" step. I suspect this is because no RPC was available for the destination chain.

Now thinking about it again, it might work when we just add the destination chain to the viem client. It would certainly be nice to also return the fill txHash. A bit busy atm but will give it a try in the next days and get back to you

typescript/agentkit/src/action-providers/across/acrossActionProvider.ts
```diff
@@ -0,0 +1,375 @@
...
+      // Get available routes
+      const routeInfo = await acrossClient.getAvailableRoutes({
+        originChainId: this.#chain.id,
+        destinationChainId: destinationChain.id,
+        originToken: inputToken,
+      });
+
+      // Select the appropriate route for native ETH or ERC20 token
+      let route;
+      for (let i = 0; i < routeInfo.length; i++) {
```
CarsonRoscoe, line 184: nit: Can be rewritten as `const route = routeInfo.find(route => route.isNative === isNative);`

typescript/agentkit/src/action-providers/across/acrossActionProvider.ts
```diff
@@ -0,0 +1,375 @@
...
+      const inputToken = tokenInfo.address as Hex;
+      const decimals = tokenInfo.decimals;
+      const inputAmount = parseUnits(args.amount, decimals);
+
+      // Check balance
+      const isNative = args.inputTokenSymbol.toUpperCase() === "ETH";
+      if (isNative) {
+        // Check native ETH balance
+        const ethBalance = await this.#publicClient.getBalance({
+          address: account.address,
```
CarsonRoscoe, line 153: nit: capture `walletProvider.getAddress() as Address` as a variable above, and use that in place of `account.address` in both places it is referenced

typescript/agentkit/src/action-providers/across/utils.ts
```diff
@@ -0,0 +1,23 @@
+/**
+ * Checks if a chain ID corresponds to a testnet network
+ *
+ * @param chainId - The blockchain network chain ID
+ * @returns true if the chain ID corresponds to a testnet, false otherwise
+ */
+export function isTestnet(chainId: number): boolean {
```
CarsonRoscoe, line 7: I have concerns about this being conflated with AgentKit supported networks, however it is valid that Across has it's own list of supported chains that will differ from ours. 

As long as the distinction is clear, I am willing to keep it. Can you rename this to something that will make that distinction clear, such as `isAcrossSupportedTestnet`?

typescript/examples/langchain-cdp-chatbot/chatbot.ts
```diff
@@ -122,6 +123,10 @@ async function initializeAgent() {
             ]
           : []),
         alloraActionProvider(),
+        acrossActionProvider({
```
CarsonRoscoe, line 126: Please remove

typescript/agentkit/src/action-providers/across/acrossActionProvider.ts
phdargen: changed

typescript/agentkit/src/action-providers/across/acrossActionProvider.ts
```diff
@@ -0,0 +1,375 @@
...
+          `Output amount has high slippage of ${actualSlippagePercentage.toFixed(2)}%, which exceeds the maximum allowed slippage of ${args.maxSplippage}%. ` +
+            `Input: ${formattedInfo.inputAmount} ${args.inputTokenSymbol}, Output: ${formattedInfo.outputAmount} ${args.inputTokenSymbol}`,
+        );
+      }
+
+      //Approve ERC20 token if needed
+      let approvalTxHash;
+      if (!isNative) {
+        const approvalAmount = quote.deposit.inputAmount;
+        approvalTxHash = await walletClient.writeContract({
```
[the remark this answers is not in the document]
phdargen: Ok I removed publicClient entirely and use walletProvider where possible.
The walletClient is still needed to simulate and execute the depositTX

typescript/agentkit/src/action-providers/across/acrossActionProvider.ts
phdargen: done

typescript/agentkit/src/action-providers/across/utils.ts
phdargen: changed to isAcrossSupportedTestnet

phdargen:
> Please add a entry into `typescript/agentkit/README.md`'s `## Action Providers` section, so this action is discoverable via readmes
> 
> Note: entries are in alphabetical order

Added entrie to README

phdargen:
Hi @CarsonRoscoe, thanks for the feedback, I implemented the changes.

I have one question about the wallets created by CdpWalletProvider and saved in wallet_data.txt when using the chatbot.ts example script. Is there an easy way to use this wallet on a different chain? It seems editing the networkId in wallet_data.txt and reimporting has no effect. You can export the private key of the cdpWallet but then importing only supports mnemonicPhrase.

If the chain cannot be easily switched, do we need a warning for wallets created with CdpWalletProvider or do not default the recipient address to sender?

typescript/agentkit/src/action-providers/across/acrossActionProvider.ts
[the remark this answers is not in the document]
CarsonRoscoe: Understood, thank you

typescript/examples/langchain-cdp-chatbot/chatbot.ts
CarsonRoscoe: Thanks for removing it! The main reason is that the example script is meant to stay minimal for new users. It should focus on basic, common actions like wallet, faucet, and ERC20. Actions which are understood by the majority of web2 AI builders and web3 builders exploring AI. Adding more advanced or third-party actions (like bridging) adds a level of domain knowledge that we don't want in the example showcases.

Also, if we accepted every new action in the example, it’d quickly get bloated and hard to maintain.

That said, what we do want is for builders who want bridging to easily discover your action, and we understand this is an area that we can improve. I'm open to discussions on where you feel we should be highlighting all the actions

CarsonRoscoe (commented):
re: Wallet question

Yes, the current version of the Cdp server wallet exports to 1 wallet per chain. The same API keys work on all the networks, however it will be a different address/private key per network.

We understand the current wallet persistence approach is lacking. We're exploring other approaches, for example in the python side, if you run `pipx run create-onchain-agent --beginnner` to generate a chatbot, [you can see here](https://github.com/coinbase/agentkit/blob/main/python/create-onchain-agent/templates/beginner/wallet.py) that we handle the wallets on a per-network basis now.

We will be refactoring all the examples shortly to use a similar approach in order to make sure that its a different wallet generated per network.

phdargen:
> re: Wallet question
> 
> Yes, the current version of the CDP server wallet exports to 1 wallet per chain. The same API keys work on all the networks, however it will generate a different address/private key per network.
> 
> We understand the current wallet persistence approach is lacking. We're exploring other approaches, for example in the python side, if you run `pipx run create-onchain-agent --beginnner` to generate a chatbot, [you can see here](https://github.com/coinbase/agentkit/blob/main/python/create-onchain-agent/templates/beginner/wallet.py) that we handle the wallets on a per-network basis now.
> 
> We will be refactoring all the examples shortly to use a similar approach in order to make sure that its a different wallet generated per network.
> 
> The temporary solution to the question regarding networks is to rename your file when you change networks. i.e. rename to`wallet_data_base_sepolia.txt` when you switch networks and generate a new wallet, and then rename back when you go back. After the refactor, this process will happen automatically for you.

OK I see, thanks for the info. Imo it would be better to have the same address on all chains.

My concern in relation to this cross-chain action is that a user sends eg ETH from Base to Ethereum with destinationAdress = senderAdress. But then after switching to Ethereum a new wallet is created and the funds deposited at destinationAdress are not accessible (within AgentKit).

But even if that happens, I suppose its possible to export the private key and then import it in MM or another wallet, so the funds would at least not be lost.

typescript/examples/langchain-cdp-chatbot/chatbot.ts
phdargen: Ok fair enough. 

Maybe some info how to add particular actions could then be added to agentkit or chatbot README as sometimes additional configs are needed. Something like this that can be easily copied and pasted if someone wants to add this action:
```
        acrossActionProvider({
          privateKey: await (await walletProvider.getWallet().getDefaultAddress()).export(),
        }),
```

typescript/.changeset/neat-emus-exist.md
```diff
@@ -0,0 +1,5 @@
+---
+"@coinbase/agentkit": minor
+---
+
+Added a new action provider to interact with Across protocol
```
CarsonRoscoe, line 5: nit: Maybe we should add a little more info so people going through the protocol have more context.

e.g. something like:
```
Added AcrossActionProvider to allow bridging tokens using the Across protocol
- `bridge_token` action to bridge native and ERC20 tokens
- `check_deposit_status` action to check the status of bridge deposits
```

phdargen:
> > OK I see, thanks for the info. Imo it would be better to have the same address on all chains.
> > My concern in relation to this cross-chain action is that a user sends eg ETH from Base to Ethereum with destinationAdress = senderAdress. But then after switching to Ethereum a new wallet is created and the funds deposited at destinationAdress are not accessible (within AgentKit).
> > But even if that happens, I suppose its possible to export the private key and then import it in MM or another wallet, so the funds would at least not be lost.
> 
> Agreed. Fyi, multi-network wallet API support is coming soon, but in the meantime, you are right that funds are safe and can be accessed through exporting the private key. You can also swap out the `CdpWalletProvider` for a `ViemWalletProvider` in the meantime to ensure your account is the same on each network.

Added warning about this and recommendation to use ViemWalletProvider to across readme

typescript/.changeset/neat-emus-exist.md
phdargen: Ok changed it

typescript/examples/langchain-cdp-chatbot/chatbot.ts
CarsonRoscoe: Good point. We'll consider the best ways to standardize this

CarsonRoscoe (approved):
LGTM!

Nicely done

PR #571, opened 2025-03-14
chore: fix nightly

0xRAG:
## Description

Nightlies are failing because the action tries to publish built artifacts (e.g. `create-onchain-agent/templates/next`). This PR updates the action to ignore `dist`, and the `create-onchain-agent` templates.

Also ignores the GitHub Release creator until we fix some issues there:
* Links to some packages are incorrect
* Unreleased changelog needs to accumulate from the changesets / changelog.d directory

## Checklist

- [ ] Added a changelog entry

### DIFF
--- a/.github/workflows/publish_nightly.yml
+++ b/.github/workflows/publish_nightly.yml
@@ -53,15 +53,19 @@ jobs:
       - name: Find NPM packages
         id: find-npm-packages
         run: |
+          cd typescript
           PACKAGES=$(find . -name "package.json" \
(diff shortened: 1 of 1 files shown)


CarsonRoscoe (approved):
Would we rather comment it out or outright delete it?

If commenting out is your preference, this LGTM

0xRAG:
> Would we rather comment it out or outright delete it?
> 
> If commenting out is your preference, this LGTM

@CarsonRoscoe I'm going to fast follow with a fix, so I'd rather just keep it

PR #735, opened 2025-05-28
feat: aws bedrock python example

CarsonRoscoe:
Continuation of #708, which adds a AWS bedrock example in Python.

This PR then migrates that to using `CdpEvmServerWalletProvider` and updates the chatbot to the new function/data file approach

### Test

image


avidreder (approved):
LGTM

PR #831, opened 2025-08-27
fix: Add hardcoded ETH price feed ID for Pyth as stop-gap solution

murrlincoln:
## Description

This PR adds a hardcoded price feed ID for ETH in the Pyth action provider as a temporary stop-gap solution.

## Changes

- Added special case handling for ETH token symbol in both Python and TypeScript implementations
- Returns hardcoded price feed ID: `0xff61491a931112ddf1bd8147cd1b641375f79f5825126d665480874634fd0ace`
- Added comments explaining this is a temporary solution until proper new API link is provided after discussions with the Pyth team

## Context

This change addresses an issue with fetching ETH price feed ID from Pyth's API. The hardcoded value ensures consistent behavior for ETH price queries while we wait for the proper API update from the Pyth team.

## Testing

- This change maintains backward compatibility
- Other token symbols continue to use the dynamic API lookup
- ETH specifically returns the hardcoded price feed ID

## Note

This is explicitly marked as a stop-gap solution in the code comments and should be revisited once the Pyth team provides the proper API endpoint.

## Files Modified

- `python/coinbase-agentkit/coinbase_agentkit/action_providers/pyth/pyth_action_provider.py`
- `typescript/agentkit/src/action-providers/pyth/pythActionProvider.ts`


phdargen:
Looks good @murrlincoln but ts tests need to be fixed.

Is this an isolated issue only for ETH price?

murrlincoln:
> Looks good @murrlincoln but ts tests need to be fixed.
> 
> Is this an isolated issue only for ETH price?

yes, ETH shows 40 cents but BTC works fine. Longer-term I think there's a change to be made to the discovery endpoint or something

CarsonRoscoe (approved):

CarsonRoscoe (approved):

PR #833, opened 2025-09-02
fix: network support for cdp wallets

CarsonRoscoe:
## Description

Fixed CDP wallet network mappings.

#### Python

Fixed network support in CDP wallet providers. Was only working on `base-sepolia` and `ethereum-sepolia`. Was failing on supported mainnets.

#### Typescript

Expanded supported CDP wallet mapping beyond just `base-sepolia` and `base`. Added `ethereum`, `ethereum-sepolia`, `polygon`, `arbitrum` and `optimism` networks.

## Checklist

- [x] Added a changelog entry

### DIFF
--- a/typescript/agentkit/src/wallet-providers/cdpEvmWalletProvider.ts
+++ b/typescript/agentkit/src/wallet-providers/cdpEvmWalletProvider.ts
@@ -21,7 +21,14 @@ import { WalletProviderWithClient, CdpWalletProviderConfig } from "./cdpShared";
 /**
  * Supported network types for CDP SDK EVM transactions
  */
-type CdpEvmNetwork = "base" | "base-sepolia" | "ethereum" | "ethereum-sepolia" | "polygon" | "arbitrum" | "optimism";
+type CdpEvmNetwork =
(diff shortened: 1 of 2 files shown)


phdargen:
Cool, since when are those networks supported? Might be good to bump the cdp-sdk version (python), seems a bit outdated

CarsonRoscoe (commented):
@phdargen Good point. [Here is the list of supported networks](https://docs.cdp.coinbase.com/get-started/supported-networks). Python side is at >1.22.0 while Typescript side is >1.36.0.

As of 1.22.0, the earlier of the two, the tagged commit is [here](https://github.com/coinbase/cdp-sdk/blob/e98a66590037c423af70dae331686670924dcf18/typescript/src/openapi-client/generated/coinbaseDeveloperPlatformAPIs.schemas.ts#L3010). Seems these networks were already supported back then. So we shouldn't have to do an update.

However, it wouldn't hurt to do an update, specifically in Python where we've fallen behind a bit. Will still bump

0xRAG (approved):

PR #977, opened 2026-02-26
Update error message for Agentkit.from without walletProvider

mattliving:
## Description

Updates Agentkit.from constructor error message when `walletProvider` is not passed to account for ALL needed config variables. 

## Tests

No tests added.


phdargen (approved):
Thanks for the fix @mattliving

PR #1054, opened 2026-03-28
fix: resolve $ref pointers from recursive Zod schemas

### LINKED ISSUE #815: Recursive Zod Schemas lead to Invalid Schema in OpenAI
### Language Affected

- [ ] Python
- [x] TypeScript

### 🐛 Describe the bug

I am using the AgentKit template, using Typescript and Langchain with a local wallet. Works out of the box. I have created my own ActionProvider, and it also works great, except when I use a recursive schema. I have created an action using this schema (greatly simplified for clarity, but it still fails):

```
export const contractMethodParamsSchema: z.ZodType<{
  [key: string]:
    | string
    | { [key: string]: any }
    | Array<any>;
}> = z.record(
  z.string(),
  z.union([
    z.string(),
    z.lazy(() => contractMethodParamsSchema),
    z.array(z.lazy(() => contractMethodParamsSchema)),
  ]),
);

export const executeSchema = z.object({
  params: contractMethodParamsSchema,
});
```
As soon as I query the model I get this error:

```
Error processing request: BadRequestError: 400 Invalid schema for function 'OneShotActionProvider_execute-contract-method-with-1shot-wallet': In context=(), object schema missing properties.
    at APIError.generate (webpack-internal:///(rsc)/./node_modules/openai/error.mjs:60:20)
    at OpenAI.makeStatusError (webpack-internal:///(rsc)/./node_modules/openai/core.mjs:333:65)
    at OpenAI.makeRequest (webpack-internal:///(rsc)/./node_modules/openai/core.mjs:377:30)
    at process.processTicksAndRejections (node:internal/process/task_queues:95:5)
    at async eval (webpack-internal:///(rsc)/./node_modules/@langchain/openai/dist/chat_models.js:1594:29)       
    at async RetryOperation.eval [as _fn] (webpack-internal:///(rsc)/./node_modules/p-retry/index.js:50:12) {    
  status: 400,
  headers: {
    'access-control-expose-headers': 'X-Request-ID',
    'alt-svc': 'h3=":443"; ma=86400',
    'cf-cache-status': 'DYNAMIC',
    'cf-ray': '96a175719f7e7c03-LAX',
    connection: 'keep-alive',
    'content-length': '307',
    'content-type': 'application/json',
    date: 'Mon, 04 Aug 2025 22:23:23 GMT',
    'openai-organization': 'uxly-software',
    'openai-processing-ms': '127',
    'openai-project': 'proj_2EjCCsonQJXknpWRniGsDyco',
    'openai-version': '2020-10-01',
    server: 'cloudflare',
    'set-cookie': '__cf_bm=fTuAa.HBiGpxnn7OfIUFGbQnbJ6LT8mbxb3w7Hp1n9M-1754346203-1.0.1.1-8H9LF9PaeTqtGpt2ru1s_96MeDqSnHymzBeti8U.MGzlir01wLWUwqydEE97HTmh.S3mpFR7NpZ5gMUj0s940fsbUT48iMRExDN_2LGfrSA; path=/; expires=Mon, 04-Aug-25 22:53:23 GMT; domain=.api.openai.com; HttpOnly; Secure; SameSite=None, _cfuvid=CiSBt45NztZkj7OT1vwCQKO60AeF2LEoMf9.5Z5E9Zk-1754346203757-0.0.1.1-604800000; path=/; domain=.api.openai.com; HttpOnly; Secure; SameSite=None', 
    'strict-transport-security': 'max-age=31536000; includeSubDomains; preload',
    'x-content-type-options': 'nosniff',
    'x-envoy-decorator-operation': 'router.openai.svc.cluster.local:5004/*',
    'x-envoy-upstream-service-time': '238',
    'x-ratelimit-limit-requests': '10000',
    'x-ratelimit-limit-tokens': '200000',
    'x-ratelimit-remaining-requests': '9999',
    'x-ratelimit-remaining-tokens': '199663',
    'x-ratelimit-reset-requests': '8.64s',
    'x-ratelimit-reset-tokens': '101ms',
    'x-request-id': 'req_9dd533995e7533c215a7b1a8a93825d4'
  },
  request_id: 'req_9dd533995e7533c215a7b1a8a93825d4',
  error: {
    message: "Invalid schema for function 'OneShotActionProvider_execute-contract-method-with-1shot-wallet': In context=(), object schema missing properties.",
    type: 'invalid_request_error',
    param: 'tools[16].function.parameters',
    code: 'invalid_function_parameters'
  },
  code: 'invalid_function_parameters',
  param: 'tools[16].function.parameters',
  type: 'invalid_request_error',
  attemptNumber: 1,
  retriesLeft: 6,
  pregelTaskId: '0002e74e-e320-5793-b0d5-f773fd759126'
}
 POST /api/agent 200 in 3904ms
```
 For reference, the action provider definition looks like this:

```
@CreateAction({
    name: "execute-contract-method-with-1shot-wallet",
    description: `...`,
    schema: executeSchema,
  })
  public async executeContractMethodWith1ShotWallet(
    args: z.infer<typeof executeSchema>,
  ): Promise<Transaction> {
```

If I comment out the two lines in the schema with the lazy recursion, then the error goes away. There seems to be an issue with generating a tool for OpenAI. I did the exact same thing (used the same schema) with the GOAT SDK, similar to AgentKit, and did not have an issue with this, so I think this is an issue inside the AgentKit layer and not a general OpenAI issue, but I'm not sure.

### Versions

Node JS Version: v20.19.1
AgentKit Version: ```
├─┬ @coinbase/agentkit-langchain@0.3.0
│ └── @coinbase/agentkit@0.9.1 deduped
├─┬ @coinbase/agentkit-vercel-ai-sdk@0.1.0
│ └── @coinbase/agentkit@0.9.1 deduped
└── @coinbase/agentkit@0.9.1
```
AgentKit langchain version: @coinbase/agentkit-langchain@0.3.0

### PR DESCRIPTION
**Pull request by 4444J99**
## Description

Fixes #815

`zodToJsonSchema()` and `z.toJSONSchema()` emit `$ref` pointers when a Zod type is referenced more than once (shared sub-schemas) or uses `z.lazy()` for recursion. OpenAI's function-calling API rejects schemas containing `$ref` with `BadRequestError: 400 Invalid schema ... object schema missing properties`.

This PR adds a `resolveJsonSchemaRefs()` utility that inlines `$ref` definitions up to a configurable depth (default: 5), replacing deeper levels with a permissive empty schema (`{}`), and strips the `$defs`/`definitions` block from the output.

**Integration approach:**

- **MCP extension**: Integrated directly into `getMcpTools()` where AgentKit controls the `z.toJSONSchema()` conversion, so ref resolution happens transparently for all MCP users.
- **LangChain / Vercel AI SDK**: These extensions pass raw Zod schemas to their respective `tool()` functions, which handle JSON Schema conversion internally. The utility is exported from `@coinbase/agentkit` so users of these extensions can apply it to their own `zodToJsonSchema()` output when needed.

**Usage for LangChain/Vercel AI SDK users with recursive schemas:**

```typescript
import { resolveJsonSchemaRefs } from "@coinbase/agentkit";
import { zodToJsonSchema } from "zod-to-json-schema";

const jsonSchema = zodToJsonSchema(myRecursiveZodSchema);
const resolved = resolveJsonSchemaRefs(jsonSchema);
```

## Tests

10 unit tests for the core utility covering:
- Simple `$ref` from `$defs` and `definitions`
- Recursive schemas with depth limiting
- Shared sub-schemas in unions (exact reproduction of issue #815)
- Output stripping of definition blocks
- Primitive passthrough, unrecognized `$ref` paths, null values
- Default max depth behavior

2 integration tests for the MCP extension:
- Existing tool/handler contract preserved
- Shared sub-schemas produce no `$ref` or `$defs` in output

```
PASS src/resolveJsonSchemaRefs.test.ts (10 tests)
PASS src/index.test.ts (2 tests)
```

Lint and format checks pass.

## Checklist

- [x] Added a changelog entry


cb-heimdall:
<table>
<tr>
<th>Requirement</th>
<th>Status</th>
<th>More Info</th>
</tr>
<tr>
<td>Reviews</td>
<td>
🟡
<code>0/1</code>
</td>
<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>
</td>
<td>1</td>
</tr>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Max</strong></td>
<td><strong>0</strong></td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>
From <code>CODEOWNERS</code>
</td>
<td>0</td>
</tr>
<tr>
<td>Global minimum</td>
<td>0</td>
</tr>
<tr>
<td><strong>Max</strong></td>
<td>
<strong>
1
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>1</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>1</strong></td>
</tr>
</table>
</details>
</td>
</tr>
</table>

4444J99:
Hi — I noticed #1023 by @mvanhorn addresses the same `$ref` resolution problem with the same approach (`resolveJsonSchemaRefs`). They opened theirs about a week before this one, so I want to acknowledge the overlap.

The main differences in this PR:
- **Depth default**: 5 vs 3 — handles deeper recursive schemas (e.g., Notion's block tree)
- **Test coverage**: 12 tests including edge cases for mixed `$defs`/`definitions`, nested unions, and MCP tool schemas
- **MCP integration**: `resolveJsonSchemaRefs` is called inside `getMcpTools()` so MCP users get transparent ref resolution
- **Changeset included**

Happy to defer to whichever PR the maintainers prefer, or to collaborate with @mvanhorn on merging the best parts of both. Just let us know which direction works.

4444J99:
Hi team — just following up on this. I noticed #1023 addresses the same recursive Zod schema issue. To help with evaluation, here's how the two approaches compare:

| | This PR (#1054) | #1023 |
|---|---|---|
| Default depth | 5 | 3 |
| Test coverage | 12 unit tests (nested refs, circular, mixed types) | — |
| MCP integration | Full `handleSendCdpActionGroup` integration | — |
| Changelog | ✓ | — |

The deeper default (5) handles real-world schemas like `TransactionInput` → `TransferInput` → `EthAmount` → nested types. Happy to adjust if 3 is preferred, but the test suite validates correctness at depth 5.

Let me know if there's anything I can do to help move this forward. Thanks!

percymcn:
Ran this against real generator output before commenting, because the resolver's stated contract is "`resolved` contains no `$ref` — safe for OpenAI function calling". The approach is right and the whole-definition case works exactly as documented; stripping `$defs` and capping depth are both reasonable for the OpenAI target. But **neither of the two schemas this PR exists for actually comes out ref-free**, including the one in the title.

## The cause

```ts
const match = node.$ref.match(/^#\/(?:\$defs|definitions)\/(.+)$/);
const def = definitions[match[1]];
```

`(.+)` is greedy across `/`, and `match[1]` is used **raw as an object key**. A JSON Pointer is a *token sequence*, not a name — so `#/definitions/S/properties/inner/properties/one` looks up a definition literally called `S/properties/inner/properties/one`, finds nothing, and takes the `if (!def) return node` branch. The `$ref` survives — and the `$defs`/`definitions` block has already been stripped by the loop below, so the output is a **dangling** pointer rather than an unresolved one.

That matters because a pointer *into* a definition is what `zodToJsonSchema` emits whenever a sub-schema is reused or recursive — the two cases named in the docstring and the title.

## Measured (this PR's function, copied verbatim; `zod@3` + `zod-to-json-schema@3.24.5`)

| input | `$ref` left in output | `openai@7.4.0` `toStrictJsonSchema` |
|---|---|---|
| reused sub-schema (`Inner.shape.one`) | `#/definitions/S/properties/inner/properties/one` | **throws** — ``Local $ref at `properties/echo` does not resolve`` |
| recursive `z.lazy` | `#/definitions/S/properties/node` | **throws** — ``Local $ref at `properties/node/properties/next` does not resolve`` |
| plain object *(control)* | none | ACCEPTED |

```ts
const Inner = z.object({ one: z.string().min(3), two: z.number() });
resolveJsonSchemaRefs(zodToJsonSchema(z.object({ inner: Inner, echo: Inner.shape.one }), "S"));
// -> { ..., echo: { $ref: "#/definitions/S/properties/inner/properties/one" } }  and no definitions block
```

`toStrictJsonSchema` is exported from `openai/lib/transform`, so this is checkable in CI with no API key — it is the same function the SDK runs before sending, and it names the surviving pointer in the error.

## Two notes for whoever evaluates this

**1. #1023 has the same line.** I checked because the thread is waiting on a choice between them: `#1023` carries the identical `/^#\/(?:\$defs|definitions)\/(.+)$/` and the same raw-key lookup, so picking either PR ships this. Not an argument for one over the other — just that the decision doesn't resolve it.

**2. The tests can't fail on it.** Every fixture in both PRs is hand-written and uses a whole-definition ref (`#/$defs/Address`, `#/$defs/TreeNode`, `#/definitions/Item`, `#/$defs/SubSchema`). None is a pointer into a definition, so both suites pass green while the motivating cases break. A fixture that discriminates has to come from `zodToJsonSchema` itself rather than be written by hand — the shape only appears once a real generator emits it.

## Fix shape

Parse the fragment as a pointer and walk it from the document root, decoding each token (`%20`) and then unescaping RFC 6901 (`~1` → `/`, `~0` → `~`) before using it as a key:

```ts
function tokens(ref: string): string[] | null {
  if (!ref.startsWith("#/")) return null;              // "#" or "#anchor" is not a pointer
  return ref.slice(2).split("/").map(t => {
    let d: string;
    try { d = decodeURIComponent(t); } catch { d = t; } // malformed % must not throw
    return d.replace(/~1/g, "/").replace(/~0/g, "~");
  });
}
```

Then resolve token-by-token from `schema` (not from the flattened `definitions` map, so `$defs` and `definitions` stay distinguishable). Two things worth keeping while you're in there:

- **`decodeURIComponent` throws** on a malformed escape such as `#/$defs/%zz`. Without the `try` that is an uncaught `URIError` out of a function whose job is to sanitise untrusted schema input.
- **Don't strip the bag when a `$ref` survived.** Today an unresolvable ref plus an unconditional strip turns a document whose pointer *did* resolve into one that doesn't. Keeping `definitions` when any `$ref` remains — or throwing — both fail closed; the current combination fails open and the failure surfaces later, at the API, as an error about a schema the caller never wrote.

Recursion also needs its identity to be the **whole pointer** rather than the first token, or two different pointers into one definition read as a cycle when they aren't.

PR #1449, opened 2026-08-15
feat(relayshield): add RelayShield action provider

nzdsf2-gif:
## Description

Adds a **RelayShield action provider** with four screening actions, each paid per call over x402 from the agent's own wallet.

| Action | Answers | Price |
|---|---|---|
| `screen_wallet` | Is this counterparty address associated with scams, exploits, drainers or sanctions? | $0.05 |
| `check_token_security` | Is this token a honeypot, mintable, blacklistable or otherwise restricted? | $0.05 |
| `check_nft_security` | Is this NFT collection fake, copied or transfer-restricted? | $0.10 |
| `screen_url` | Is this link phishing or malware? | $0.05 |

`screen_wallet` detects the chain from the address format and covers EVM, Solana, TON and Bitcoin, so the agent never has to ask the user which chain an address is on.

## Why no API key

The endpoints answer with an HTTP 402 challenge and the agent's wallet settles in USDC in the same request cycle. There is no account to create and no key to store or rotate, which is the only shape that works when the buyer is software. This reuses the same `x402Client` / `wrapFetchWithPayment` / `registerExact*Scheme` pattern as the existing `x402` action provider rather than introducing a second payment path.

`supportsNetwork` returns true for `base-mainnet` and `solana-mainnet`. That reflects where a **payment** can settle, not which chains can be screened, and the README says so explicitly to avoid the obvious misreading.

## A failed check is never reported as a clean one

If the endpoint errors or the payment fails, the action returns a message stating plainly that the check did not complete and the item should be treated as unverified. A screening tool that looks like it found nothing when it actually failed is worse than one that admits it could not answer, and there are tests covering both the non-200 and the thrown-error paths.

The action descriptions also instruct the model to report a clean result as "nothing currently known against it" rather than as a safety guarantee.

## Testing

- `pnpm exec jest src/action-providers/relayshield` — **9 tests, all passing**, covering network support, argument-to-body mapping for each endpoint, a HIGH verdict passing through unchanged, and both failure paths.
- `pnpm run build` — clean.
- `pnpm run lint` — clean.
- `pnpm run format` — clean.
- Full suite re-run to check for regressions: **61 suites, 873 tests, all passing.**

**Not tested:** a live end-to-end paid call against a funded wallet. The tests mock `wrapFetchWithPayment`, so the payment path itself is exercised only through the same interface the `x402` provider uses. Flagging that rather than implying broader coverage than exists.

The endpoints themselves were verified live before submitting: all four return a well-formed 402 challenge with USDC payment options on both Base and Solana.

## Checklist

- [x] Changeset added (`patch`, per the contributing guide's guidance for new action providers)
- [x] README for the action provider
- [x] Export added to `action-providers/index.ts`, in alphabetical position
- [x] Generated with `generate-action-provider` rather than hand-rolled

### DIFF
--- a/typescript/.changeset/olive-donkeys-shave.md
+++ b/typescript/.changeset/olive-donkeys-shave.md
@@ -1,0 +1,5 @@
+---
+"@coinbase/agentkit": patch
+---
+
+Added a new action provider to screen counterparties and content with RelayShield, paid per call over x402
--- a/typescript/agentkit/src/action-providers/index.ts
+++ b/typescript/agentkit/src/action-providers/index.ts
@@ -23,6 +23,7 @@ export * from "./pyth";
 export * from "./moonwell";
 export * from "./morpho";
 export * from "./opensea";
+export * from "./relayshield";
 export * from "./spl";
 export * from "./superfluid";
 export * from "./sushi";
--- a/typescript/agentkit/src/action-providers/relayshield/README.md
+++ b/typescript/agentkit/src/action-providers/relayshield/README.md
@@ -1,0 +1,65 @@
+# RelayShield Action Provider
+
+Counterparty and content screening for agents, paid per call over x402.
+
+An agent about to send funds to an address, swap into a token, buy an NFT, or follow a link
+has a question it cannot answer from onchain data alone: is the thing on the other side known
+to be malicious? This provider answers that question, and the agent pays for the answer out of
+its own wallet.
+
+## Why this needs no API key
+
+Every action calls a RelayShield pay-as-you-go endpoint, which replies with an HTTP 402
+challenge. The agent's wallet settles the payment in USDC and the call proceeds in the same
+request cycle. There is no account to create, no key to store and no key to rotate, which is
+the only shape that works when the buyer is software.
+
+## Actions
+
+| Action | Answers | Price |
+|---|---|---|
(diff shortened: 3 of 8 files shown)


cb-heimdall:
<table>
<tr>
<th>Requirement</th>
<th>Status</th>
<th>More Info</th>
</tr>
<tr>
<td>Reviews</td>
<td>
🟡
<code>0/2</code>
</td>
<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>
</td>
<td>1</td>
</tr>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Max</strong></td>
<td><strong>0</strong></td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>
From <code>CODEOWNERS</code>
</td>
<td>0</td>
</tr>
<tr>
<td>Global minimum</td>
<td>0</td>
</tr>
<tr>
<td><strong>Max</strong></td>
<td>
<strong>
1
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>1</td>
</tr>
<tr>
<td>1</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>2</strong></td>
</tr>
</table>
</details>
</td>
</tr>
</table>

